PR #2, opened 2021-01-08
Add RSR and Ogrs calls to the community-api

nickmcmahon01:
Code to call the new endpoints in community-api for RSR and OGRS

- new webclient methods to call the new endpoints
- very thin service layer to extract the values from the methods
- add the service layer calls to the calculator in the right place
- tests tests tests.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/enums/ComplexityFactor.kt
```diff
@@ -12,7 +12,7 @@ enum class ComplexityFactor(val registerCode: String) {
   PUBLIC_INTEREST("RPIR"),
   ADULT_AT_RISK("RVAD"),
   STREET_GANGS("STRG"),
-  TERRORISM("TACT");
+  TERRORISM("RTAO");
```
nickmcmahon01, line 15: unrelated bug: RTAO, not TACT is the code referenced in the ticket, I misread it at the time.

.circleci/config.yml
```diff
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
+  hmpps: ministryofjustice/hmpps@2.2
```
nickmcmahon01, line 4: We need the new plugin for helm, but the new plugin introduced ktlint too so lots of what feel like unimportant linting changes ahead!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -15,13 +16,30 @@ class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val we
...
       .retrieve()
       .bodyToMono(CommunityApiRegistrationsDto::class.java)
       .block()?.registrations ?: listOf()
   }
+
+  @Cacheable(value = ["deliusAssessment"], key = "{ #crn }")
+  fun getAssessments(crn: String): DeliusAssessmentsDto? {
+    return webClient
+      .get()
+      .uri("/offenders/crn/$crn/registrations")
```
markreesmoj, line 29: Should this not end with `/assessments` ?

nickmcmahon01: It would work better wouldn't it 👍

markreesmoj (approved):
✅

PR #3, opened 2021-01-15
Add the new complexity factors from oasys

nickmcmahon01:


nickmcmahon01:
PR to check for two new Oasys complexity factors, and return 1 point if either (or both) of them are scored non-zero.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/TierCalculation.kt
```diff
@@ -146,16 +145,34 @@ class TierCalculation {
...
+
+    private fun getAssessmentComplexityPoints(factors: Map<AssessmentComplexityFactor, String?>): Int {
+
+      val parentingAnswer = factors.getOrDefault(AssessmentComplexityFactor.PARENTING_RESPONSIBILITIES, "N")
+      val parenting = if (isYes(parentingAnswer)) 1 else 0
+
+      // We dont take the cumulative score, just '1' if at least one of these two is present
+      val impulsivityAnswer = factors[AssessmentComplexityFactor.IMPULSIVITY]?.toInt() ?: 0
+      val temperControlAnswer = factors.getOrDefault(AssessmentComplexityFactor.TEMPER_CONTROL, "0")?.toInt() ?: 0
+      val selfControl = if (impulsivityAnswer.plus(temperControlAnswer) > 1) 1 else 0
```
markreesmoj, line 168: Should this be > 0 rather than > 1?
" if they have one or the other or both"

markreesmoj (commented):
Not read all of it, and I don't speak Kotlin yet, but just one observation

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/TierCalculation.kt
nickmcmahon01: it should be, good spot, its because in the tests I've used 1 and 2 as the scores, would have caught it if i'd used 0 and 1

markreesmoj (approved):
👍

PR #4, opened 2021-01-15
restrict gender specific complexity factors

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/TierCalculation.kt
```diff
@@ -145,29 +145,46 @@ class TierCalculation {
...
+        val genericScore = complexityFactors.distinct().count() * multiplicationFactor
+
+        when (isFemale) {
+            true -> {
+                genericScore.plus(getAssessmentComplexityPoints(assessmentComplexityFactors) * 2)
+            }
+            else -> {
+              when {
+                complexityFactors.contains(ComplexityFactor.IOM_NOMINAL) -> {
+                  genericScore - multiplicationFactor
```
markreesmoj, line 164: Again, knowing nothing of Kotlin so please ignore if irrelevant. If there is a .plus function (159) should this use .minus?

Also a variable called multiplicationFactor could perhaps have a better name if it's being subtracted rather than multiplied (unless it is a domain specific name of course)

src/test/kotlin/domain/TierCalculationResultCalculationTest.kt
```diff
@@ -43,7 +43,7 @@ internal class TierCalculationResultCalculationTest {
         assessmentComplexityFactors = mapOf()
       )
 
-      val tier = calculator.calculateTier(protectScores, changeScores)
+      val tier = calculator.calculateTier(protectScores, changeScores, true)
```
markreesmoj, line 46: Looks like all the tests in this file are for female offenders. At the moment I think this is only indicated by the `true` which has been added to the arguments.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/TierCalculation.kt
nickmcmahon01: No you're right, good catch, I should use .minus() to be consistent. I'll also rename multiplicationFactior and put a comment, I should also use it on line 159 where the '2' is.

src/test/kotlin/domain/TierCalculationResultCalculationTest.kt
nickmcmahon01: Yep I've got a few 'male' tests to add

PR #5, opened 2021-01-19
Remove the extra layer of abstraction from when we calculated the old…

nickmcmahon01:
… and new tiers, moving the new calculation business logic into the service layer


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/enums/ProtectLevel.kt
```diff
@@ -1,6 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppstier.domain.enums
 
-enum class ProtectScore {
+enum class ProtectLevel {
```
nickmcmahon01, line 3: Domain - we get assessment SCORES, which map to threshhold POINTS which returns a LEVEL

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/enums/TierMatchCriteria.kt
```diff
@@ -1,23 +0,0 @@
-package uk.gov.justice.digital.hmpps.hmppstier.domain.enums
```
nickmcmahon01, line 1: This was an invented requirement, I think we're much better off either saving the scores used to calculate the tier with the score or doing nothing

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -67,10 +58,125 @@ class TierCalculationService(
     } else {
       log.info("Found latest tier calculation for $crn")
     }
-
     return calculation
   }
```
nickmcmahon01, line 63: Just moved the code from a component into the service as used to pass 'OldCalcuation' and 'NewCalculation' in, but now we just have one let of BL it doesn't make sense to have the extra complexity.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -30,27 +34,13 @@ class TierCalculationService(
...
-      assessmentComplexityFactors = if (isFemale) assessmentApiDataService.getAssessmentComplexityAnswers(crn) else emptyMap()
-    )
-
-    val changeScores = ChangeScores(
-      crn = crn,
-      ogrsScore = communityApiDataService.getOGRS(crn),
-      need = assessmentApiDataService.getAssessmentNeeds(crn)
-    )
+    val protectLevel = calculateProtectLevel(crn)
+    val changeLevel = calculateChangeLevel(crn)
```
nickmcmahon01, line 38: i've left this as a seperate val because we're just going to change it with the mandate for change work

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -67,10 +57,122 @@ class TierCalculationService(
...
+
+  private fun getMappaPoints(crn: String): Int {
+    return when (communityApiDataService.getMappa(crn)) {
+      Mappa.M2, Mappa.M3 -> 30
+      Mappa.M1 -> 5
+      else -> 0
+    }
+  }
+
+  private fun getComplexityPoints(crn: String): Int {
```
nickmcmahon01, line 123: Not sure on the readability here, the other WHEN()s are much better

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/dto/TierDto.kt
```diff
@@ -4,36 +4,37 @@ import com.fasterxml.jackson.annotation.JsonCreator
...
-  @JsonProperty("protectScore")
-  val protectScore: Int?,
+  @ApiModelProperty(value = "Protect Points", example = "17")
+  @JsonProperty("protectPoints")
+  val protectPoints: Int?,
 
-  @ApiModelProperty(value = "Change Tier", example = "2")
-  @JsonProperty("changeTier")
-  val changeTier: ChangeScore,
+  @ApiModelProperty(value = "Change Level", example = "2")
```
markreesmoj, line 22: The example looks like it should be "TWO" rather than "2" according to the ChangeLevel enum

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
markreesmoj: as long as it works and the Test coverage is there for all the branches

markreesmoj (approved):
Looks good to me

PR #6, opened 2021-01-22
Check for breach and recall

nickmcmahon01:


src/test/kotlin/domain/enums/ComplexityFactorTest.kt
```diff
@@ -94,10 +94,5 @@ class ComplexityFactorTest {
     fun `It should match case insensitive mixed`() {
       assertThat(ComplexityFactor.from("RtAO")).isEqualTo(ComplexityFactor.TERRORISM)
     }
-
```
nickmcmahon01, line 97: This is a duplicate test.

markreesmoj (approved):
👍

PR #19, opened 2021-02-04
MAS-246 only calculate needs if custodial sentence or unpaid work or restrictive requirements

markreesmoj:
Always calculate the needs axis for current sentences unless it is a non custodial sentence AND has restrictive requirements OR unpaid work

I think this is functionally complete

We could make fewer calls to community-api


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/enums/ChangeLevel.kt
```diff
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppstier.domain.enums
 
 enum class ChangeLevel {
+  ZERO,
```
nickmcmahon01, line 4: I had this and removed it previously, I guess its semantics, but is there a difference between a 0 and a null, and what do we want them to mean?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculator.kt
```diff
@@ -11,14 +11,35 @@ class ChangeLevelCalculator(
...
-      points in 10..19 -> ChangeLevel.TWO
-      else -> ChangeLevel.ONE
+    if (shouldCalculateChangeLevel(crn)) {
+
+      val points = getOasysNeedsPoints(crn).plus(getOgrsPoints(crn))
+      val tier = calculateTier(points)
+
+      return TierLevel(tier, points)
+    }
+    return TierLevel(ChangeLevel.ZERO, 0)
```
nickmcmahon01, line 21: Same here, have we calculated a tier level of 0 or have we not calculated a tier level.
I do have a recollection of siding on '0' for the int score here, but in context I wonder if the object should be null?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculator.kt
```diff
@@ -11,14 +11,35 @@ class ChangeLevelCalculator(
...
+  }
+
+  private fun shouldCalculateChangeLevel(crn: String): Boolean {
+    val isCurrentCustodial = communityApiDataService.isCurrentCustodialSentence(crn)
```
nickmcmahon01, line 25: Not blocking, but would this be clearer using a WHEN?

```
return when {
dsffdsf -> true
dsfdsfds -> method()
else -> false
}
```

nickmcmahon01 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -60,8 +60,27 @@ class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val we
...
+    return webClient
+      .get()
+      .uri("offenders/crn/$crn/convictions/$convictionId/requirements")
+      .retrieve()
+      .bodyToMono(Requirements::class.java)
+      .block()?.requirements ?: listOf()
+  }
 }
 
+data class Requirements @JsonCreator constructor(
```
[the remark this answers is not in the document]
markreesmoj: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -82,7 +101,23 @@ data class Conviction @JsonCreator constructor(
 
 data class Sentence @JsonCreator constructor(
   @JsonProperty("terminationDate")
-  var terminationDate: LocalDate?
+  var terminationDate: LocalDate?,
+
+  @JsonProperty("sentenceType")
+  val sentenceType: SentenceType,
+
+  @JsonProperty
```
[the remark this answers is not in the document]
markreesmoj: It is - I mean they all have them but the default is to use the field name, and they are mostly the same as the field name. Added anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculator.kt
markreesmoj: It's definitely more Kotlin. I should probably read a book on it or something

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculator.kt
markreesmoj: I slacked you a conversation I had with Ed about this

PR #20, opened 2021-02-05
Add coverage

markreesmoj:
This adds the jacoco code coverage plugin. 

It gives the ability to generate HTML coverage reports on demand and set a minimum coverage threshold. The reports are a bit more detailed than what intellij provides especially in the area of branch coverage

It does not fail the build if coverage drops below the threshold. If we did want to start doing that, it would be a question of making `jacocoTestCoverageVerification` run every time we run check

It turned out to be pretty easy to do this - probably an hour in total. The biggest problem was intellij complaining about the syntax.

Anyway here it is, we can either merge it or ditch it.


nickmcmahon01 (approved):
with it being generated on circle-ci can you get access to the report easily?

markreesmoj:
The check job now runs the coverage report and saves it in CircleCI.
Also upgraded the gradle spring boot plugin to fix the overnight OWASP build
This meant I had to fix the Spring config. I went for a quick hack by enabling legacy behaviour rather than tackling the root cause as we're only using it in tests.
Also worth noting - app insights agent will be v3.0. I don't think that affects things right now but I could be wrong so I'm mentioning it in case.

PR #23, opened 2021-02-10
Move IOM Nomial from protect to change.

nickmcmahon01:
Move IOM Nomial from protect to change.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
@@ -65,6 +65,7 @@ class ProtectLevelCalculator(
 
   private fun getComplexityPoints(crn: String): Int {
     val complexityFactors = communityApiDataService.getComplexityFactors(crn).distinct().filter {
+      // Filter out IOM it is not used in this calculation but is used in the change level calculation
       it != ComplexityFactor.IOM_NOMINAL
     }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -64,33 +64,30 @@ class ProtectLevelCalculator(
...
   }
 
   private fun getAssessmentComplexityPoints(crn: String): Int {
     return assessmentApiDataService.getAssessmentComplexityAnswers(crn).let {
       val parenting = when {
-        isYes(it[AssessmentComplexityFactor.PARENTING_RESPONSIBILITIES]) -> 1
+        isYes(it[AssessmentComplexityFactor.PARENTING_RESPONSIBILITIES]) -> 2
         else -> 0
       }
       // We dont take the cumulative score, just '1' if at least one of these two is present
```
markreesmoj, line 88: Looks like this comment has gone stale?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -99,7 +96,7 @@ class ProtectLevelCalculator(
 
   private fun getBreachRecallComplexityPoints(crn: String): Int {
     return when {
-      communityApiDataService.hasBreachedConvictions(crn) -> 1
+      communityApiDataService.hasBreachedConvictions(crn) -> 2
```
markreesmoj, line 99: It feels like there's a general idea that everything needs to be multiplied by 2, but that it's being done in quite a few different places

markreesmoj (approved):

PR #25, opened 2021-02-11
MAS-311 handle an empty registrations request

markreesmoj:
I took the opportunity to write something pretty close to an end to end integration test - calling the listener and checking it writes the calculated tier to the repository


nickmcmahon01 (approved):
Good spot because we already do the ? listof() anyway...!

PR #26, opened 2021-02-11
speed up deployment in line with root project

markreesmoj:


nickmcmahon01 (approved):
This should be fine, we'll have to check we have the CPU/Memory allocations to allow us to start up in (30 + (10 * 55)) seconds...only one way to find out!
It sounds a lot but when you're applying flyway scripts etc it can start to be problematic

PR #29, opened 2021-02-12
Call community-api with the tier update

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -26,6 +26,6 @@ class TierCalculationController(private val tierCalculationService: TierCalculat
   )
   @GetMapping("crn/{crn}/tier")
   fun getNotifications(@PathVariable(required = true) crn: String): ResponseEntity<TierDto> {
-    return ResponseEntity.ok(tierCalculationService.getTierByCrn(crn))
+    return ResponseEntity.ok(tierCalculationService.getOrCalculateTierByCrn(crn))
```
markreesmoj, line 29: This is just renaming the method - it does the same as it always did

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/dto/CalculationResultDto.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.hmppstier.dto
+
+data class CalculationResultDto constructor(
+  val tierDto: TierDto?
+)
```
markreesmoj, line 5: This is not very nice. The requirement is to check whether the calculation has changed, and the difficulty is when there is no previous calculation. I haven't found an elegant way of doing this and I didn't want to change the TierDto

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/MockedEndpointsTestBase.kt
```diff
@@ -31,6 +34,13 @@ abstract class MockedEndpointsTestBase : IntegrationTestBase() {
     mockAssessmentApiServer.stop()
   }
 
+  fun calculationMessage(crn: String): String {
+    val validMessage: String =
+      Files.readString(Paths.get("src/test/resources/fixtures/sqs/tier-calculation-event.json")).replace("X373878", crn)
+
+    return validMessage
+  }
+
```
markreesmoj, line 43: making it easier to write outside-in tests where we just call the listener

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/TierCalculationTest.kt
```diff
@@ -1,204 +1,218 @@
...
-import uk.gov.justice.digital.hmpps.hmppstier.service.TierCalculationService
+import uk.gov.justice.digital.hmpps.hmppstier.service.TierCalculationRequiredEventListener
 
-@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@TestInstance(PER_CLASS)
 class TierCalculationTest : MockedEndpointsTestBase() {
 
   @Autowired
-  lateinit var service: TierCalculationService
+  lateinit var listener: TierCalculationRequiredEventListener
```
markreesmoj, line 16: Now more of an outside-in test where we call the listener and check that the update happens as expected. This makes it a bit easier to refactor the service - changing the return types for example made a real mess of this

If/when we get requirements for the controller, this kind of test would be easier to write against that.

markreesmoj (commented):
reviewed by accident

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/dto/CalculationResultDto.kt
nickmcmahon01: Is this not a case of pulling the existing tier out of the database, calculating the new one, and comaparing
then 
`
old = getOld(crn)
new = calculateNew(crn)
if (old != new) { writebacktodelius() }`

markreesmoj: yes, as long as the old one is there. I didn't want to expose the DB entity in the listener though. 

Ultimately there's quite a lot going on in the listener, probably need to push some of it down

markreesmoj: The other complication is supporting the controller, for which we currently have no specs. What the listener needs in order to call the updater does not have a lot in common with what the controller returns

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationServiceTest.kt
```diff
@@ -162,7 +165,7 @@ internal class TierCalculationServiceTest {
       val tier = service.calculateTierForCrn(crn)
       standardVerify()
 
-      assertThat(tier.data.protect.points).isEqualTo(20)
```
markreesmoj, line 165: Every time I change anything, I seem to have to spend 20 minutes in this test despite none of the functionality changing

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -69,6 +71,13 @@ class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val we
       .bodyToMono(Requirements::class.java)
       .block()?.requirements ?: listOf()
   }
+
+  fun updateTier(tierDto: TierDto, crn: String): ResponseEntity<Void>? {
+    return webClient
+      .post()
+      .uri("offenders/crn/$crn/tier/${tierDto.protectLevel}${tierDto.changeLevel.ordinal}")
+      .retrieve().toBodilessEntity().block()
```
markreesmoj, line 79: This endpoint does return the offender but I suppose we aren't interested in that

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -17,15 +18,27 @@ class TierCalculationService(
...
       log.info("Returned tier for $crn")
     }
   }
 
-  fun calculateTierForCrn(crn: String): TierCalculationEntity {
+  fun calculateTierForCrn(crn: String): CalculationResultDto {
+    val existingCalculation = getLatestTierCalculation(crn)
+    val newTier = calculateTier(crn)
+    val isUpdated: Boolean
+    if (existingCalculation == null) {
```
markreesmoj, line 32: Kotlin magic needed here

nickmcmahon01 (approved):

PR #34, opened 2021-02-15
Default change level to '2' if the Assessment is too old.

nickmcmahon01:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationServiceTest.kt
```diff
@@ -85,6 +85,7 @@ internal class TierCalculationServiceTest {
 
     @Test
     fun `Should Call Collaborators Test - Female - Existing not found`() {
+      every { assessmentApiDataService.isAssessmentRecent(crn) } returns true
```
markreesmoj, line 88: this test  has about 200 lines of change in all the recent PRs

markreesmoj (approved):
👍

PR #38, opened 2021-02-18
Preprod and Prod environment values 

nickmcmahon01:
values taken from court-case-service


markreesmoj:
Looks OK to me. Is it worth adding 

FLAGS_ENABLEDELIUSTIERUPDATES: false

markreesmoj (approved):
👍

PR #45, opened 2021-02-22
Update our interactions with Assessment-API

nickmcmahon01:
* Use assessment summary instead of latest assessment endpoint
* pass the section for the question answers

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiDataService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiDataService.kt
@@ -23,7 +23,7 @@ class AssessmentApiDataService(
       // Will be not null because it is completed
       it.completed!!.toLocalDate().isAfter(cutOff)
     }.also {
-      log.debug("Assessment is recent for $crn: $it ")
+      log.debug("Found a valid Assessment $crn: $it ")
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiDataService.kt
```diff
@@ -16,15 +18,17 @@ class AssessmentApiDataService(
 ) {
 
   fun isAssessmentRecent(crn: String): Boolean {
-    return assessmentApiClient.getLatestAssessment(crn).let {
+    return getLatestCompletedAssessment(crn).let {
       val cutOff = LocalDate.now(clock).minusWeeks(55).minusDays(1)
-      it.completed.toLocalDate().isAfter(cutOff)
+      // Will be not null because it is completed
+      it.completed!!.toLocalDate().isAfter(cutOff)
```
markreesmoj, line 24: if it's never null here (although it looks like it can be when retrieved from the API), it might be an idea to introduce another type with a non-nullable when processing these assessments. Seems a  shame to have to do this otherwise

This assessments API gives back some very horrible data

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/CommunityApiDataService.kt
```diff
@@ -62,9 +62,6 @@ class CommunityApiDataService(private val communityApiClient: CommunityApiClient
 
   fun isFemaleOffender(crn: String): Boolean {
     return getOffenderGender(crn).equals("Female", true)
```
markreesmoj, line 64: I had to look this up. The true is for ignore case, interesting flag. For readability I would prefer 
getOffenderGender(crn).toLowerCase().equals("female")

Might even lowercase it in getOffenderGender (great name) - I can't imagine we will ever need it to be in whatever original mess it is in feMALe

markreesmoj (approved):

PR #48, opened 2021-02-24
Start to move the fetching of data out so that we only do it once at …

nickmcmahon01:
…the beginning.


markreesmoj:
it's big. I'm sure it's fine!

markreesmoj:
This is too big for me to review effectively. If the tests pass I'm saying it's fine.

If there's ever an opportunity to move tests higher up the application away from testing individial classes and towards testing the listener or the service, I've found that these kind of refactors (where nothing functionally changes) become easier as you don't break lots of tests when moving functionality between classes.

PR #55, opened 2021-03-03
Nsis response can be empty

markreesmoj:


nickmcmahon01 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -150,7 +150,7 @@ data class Requirement @JsonCreator constructor(
 
 private data class NsiWrapper @JsonCreator constructor(
   @JsonProperty("convictionId")
```
nickmcmahon01, line 152: This convictionId looks like a bug, does this work most of the time?!

nickmcmahon01 (changes requested):

nickmcmahon01 (approved):

PR #56, opened 2021-03-03
minor logging changes

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -68,7 +67,7 @@ class TierCalculationService(
     )
 
     return tierCalculationRepository.save(calculation).also {
-      log.info("Calculated tier for $crn")
+      log.info("Calculated tier for $crn ${calculation.data}")
```
nickmcmahon01, line 70: I'm not sure we can log this at info because it has the CRN and the tier in the same message. 
Anything logged at Info will end up in Application Insights, but Debug should only be output on dev machines

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -66,8 +66,7 @@ class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val we
   fun updateTier(tier: String, crn: String): ResponseEntity<Void>? {
     return updateTierCall(tier, crn)
       .also {
-        log.info("Updated Tier for $crn")
-        log.debug("Body: $tier for $crn")
+        log.info("Updated Tier for $crn with $tier")
```
nickmcmahon01, line 69: As below:

`I'm not sure we can log this at info because it has the CRN and the tier in the same message.
Anything logged at Info will end up in Application Insights, but Debug should only be output on dev machines`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
markreesmoj: Taking this to its logical conclusion, we shouldn't log the calculated tier anywhere at Info then really, as it's going to be easily tied back to the CRN in ApplicationInsights based on the sequence of events

PR #57, opened 2021-03-04
Add calculationID to the entity -> DTO

nickmcmahon01:


src/main/resources/db/migration/V2_0__RECREATE_TIER_CALCULATION.sql
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS TIER_CALCULATION;
```
nickmcmahon01, line 1: obviously here we'll lose data,
We can clear all the databases before we go live and consolidate these scripts.

markreesmoj: If we get this in prod before testing starts it will be fine I think.

markreesmoj:
Maybe some of the lower level tests can be replaced by tests of the endpoint, once written?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -28,9 +29,16 @@ class TierCalculationController(private val tierCalculationService: TierCalculat
       ApiResponse(code = 404, message = "Result Not Found")
     ]
   )
+
   @GetMapping("crn/{crn}/tier")
-  fun getNotifications(@PathVariable(required = true) crn: String): ResponseEntity<TierDto> {
-    return ResponseEntity.ok(tierCalculationService.getTierByCrn(crn))
+  fun getLatestTierCalculation(@PathVariable(required = true) crn: String): ResponseEntity<TierDto> {
```
markreesmoj, line 34: Do we need this endpoint now?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -28,9 +29,16 @@ class TierCalculationController(private val tierCalculationService: TierCalculat
...
+    return ResponseEntity.ok(tierCalculationService.getLatestTierByCrn(crn))
+      ?: throw EntityNotFoundException("Tier Result Not Found for $crn")
+  }
+
+  @GetMapping("crn/{crn}/tier/{calculationId}")
+  fun getTierCalculationById(@PathVariable(required = true) crn: String, @PathVariable(required = true) calculationId: UUID): ResponseEntity<TierDto> {
+    return ResponseEntity.ok(tierCalculationService.getTierByCalculationId(crn, calculationId))
       ?: throw EntityNotFoundException("Tier Result Not Found for $crn")
```
markreesmoj, line 42: should we include the UUID in this message?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/repository/TierCalculationRepositoryTest.kt
```diff
@@ -57,8 +58,55 @@ class TierCalculationRepositoryTest(
...
+      val calculationId = UUID.randomUUID()
+      val created = LocalDateTime.now()
+      val firstTierCalculation = TierCalculationEntity(crn = crn, created = created, data = data, uuid = calculationId)
+
+      repository.save(firstTierCalculation)
+
+      val calculation = repository.findByCrnAndUuid(crn, calculationId)
+      assertThat(calculation).isNotNull
+      assertThat(calculation?.crn).isEqualTo(firstTierCalculation.crn)
+      assertThat(calculation?.created).isEqualToIgnoringNanos(firstTierCalculation.created)
```
markreesmoj, line 91: 😮 
What a remarkable matcher

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/repository/TierCalculationRepositoryTest.kt
```diff
@@ -57,8 +58,55 @@ class TierCalculationRepositoryTest(
     fun `Should return latest calculation when multiple`() {
 
       val created = LocalDateTime.now()
-      val firstTierCalculation = TierCalculationEntity(crn = crn, created = created, data = data)
-      val secondTierCalculation = TierCalculationEntity(crn = "$crn 2", created = created.minusSeconds(1), data = data)
+      val firstTierCalculation = TierCalculationEntity(crn = crn, created = created, data = data, uuid = UUID.randomUUID())
+      val secondTierCalculation = TierCalculationEntity(crn = "$crn 2", created = created.minusSeconds(1), data = data, uuid = UUID.randomUUID())
```
markreesmoj, line 62: I might be misreading this but it looks like the repository has calculations with different CRNs so findFirst will always return the one which matches the crn

markreesmoj (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
nickmcmahon01: I think we need to clarify this. I can't think of a specific use case, but we master the data now so need to make it available to other clients?

markreesmoj: Feels like we are doing half a job here. If there is a need for it we need to get the access restrictions and documentation etc right, and test it.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/repository/TierCalculationRepositoryTest.kt
nickmcmahon01: That's a really good point, I'll check the git blame. 🤡

PR #62, opened 2021-03-08
MAS-347 handle financial management and emotional needs with no score

markreesmoj:


src/test/resources/fixtures/community-api/assessments.json
```diff
@@ -1,4 +1,4 @@
 {
   "rsrScore": 23.0,
-  "ogrsScore": 16
+  "ogrsScore": 21
```
markreesmoj, line 3: I just changed this so it made the needs work testable via the entry point. If this is not a valid OGRS score I will have to find another way to score 2 points

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/enums/Need.kt
```diff
@@ -8,5 +8,7 @@ enum class Need(val weighting: Int) {
   DRUG_MISUSE(1),
   ALCOHOL_MISUSE(1),
   THINKING_AND_BEHAVIOUR(2),
-  ATTITUDES(2)
+  ATTITUDES(2),
+  FINANCIAL_MANAGEMENT_AND_INCOME(0),
+  EMOTIONAL_WELL_BEING(0)
 }
```
nickmcmahon01, line 14: This is fine, if there are many more, we should change the client to return String and try to parse the results into a  Need in the service layer and then filter the null ones out like we do for other calls. This will  break if there are new Needs added.

src/test/resources/fixtures/community-api/assessments.json
nickmcmahon01: Should be fine.

nickmcmahon01 (approved):

PR #65, opened 2021-03-12
Remove the autoscaler

nickmcmahon01:


markreesmoj (approved):
As long as it doesn't cost us a significant amount to have the extra pods when we're not going to be using them very often this is fine

PR #67, opened 2021-03-18
Reduce the message quantity to avoid the rate limits, and scale the a…

nickmcmahon01:
…pplication to 2 instead 1.

So we do 1 message * 2 instances instead of 2 messages * 1 instance for more resiliency.


markreesmoj (approved):
👍

PR #69, opened 2021-03-20
Swap communityAPI call for SNS

nickmcmahon01:


helm_deploy/values-dev.yaml
```diff
@@ -22,7 +22,6 @@ env:
   OAUTH_ROOT_URL: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
   COMMUNITY_ENDPOINT_URL: https://community-api-secure.test.delius.probation.hmpps.dsd.io/secure
   ASSESSMENT_ENDPOINT_URL: https://offender-dev.aks-dev-1.studio-hosting.service.justice.gov.uk
-  FLAGS_ENABLEDELIUSTIERUPDATES: true
```
markreesmoj, line 25: Have we got any way of disabling the community-api writes in production (or elsewhere)?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/AwsConfiguration.kt
```diff
@@ -17,19 +19,31 @@ import org.springframework.context.annotation.Primary
 @Configuration
 @ConditionalOnProperty(name = ["offender-events.sqs-provider"], havingValue = "aws")
 class AwsConfiguration(
-  @Value("\${aws.access-key-id}") val accessKeyId: String,
-  @Value("\${aws.secret-access-key}") val secretKey: String,
-  @Value("\${aws.region}") val region: String
+  @Value("\${aws.offender.access-key-id}") val offenderEventsAccessKeyId: String,
```
markreesmoj, line 22: AWS_OFFENDER_EVENTS_ACCESS_KEY is the nname of this variable according to _envs.tpl

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/AwsConfiguration.kt
```diff
@@ -17,19 +19,31 @@ import org.springframework.context.annotation.Primary
...
+    val credentials: AWSCredentials = BasicAWSCredentials(offenderEventsAccessKeyId, offenderEventsSecretKey)
     return AmazonSQSAsyncClientBuilder
       .standard()
-      .withRegion(region)
+      .withRegion(offenderEventsRegion)
+      .withCredentials(AWSStaticCredentialsProvider(credentials)).build()
+  }
 
+  @Primary
+  @Bean(name = ["hmmpsDomainEvents"])
```
markreesmoj, line 41: hmpps rather than hmmps

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/SuccessUpdater.kt
```diff
@@ -1,12 +1,33 @@
...
+
+  fun update(crn: String, calculationId: UUID) {
+    val tierChangeEvent = TierChangeEvent(EventType.HMPPS_TIER_CALCULATION_COMPLETE, crn, calculationId)
+    val message = Message(gson.toJson(tierChangeEvent), calculationId.toString())
+    amazonSNS.publish(topic, gson.toJson(message))
   }
 }
+
+private data class TierChangeEvent(
+  val eventType: EventType?,
```
markreesmoj, line 24: isn't this always populated?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/SuccessUpdater.kt
```diff
@@ -1,12 +1,33 @@
...
-    client.updateTier("${tierDto.protectLevel.value}${tierDto.changeLevel.value}", crn)
+class SuccessUpdater(
+  private val amazonSNS: AmazonSNSAsync,
+  @Value("\${hmpps-events.topic}")private val topic: String,
+  private val gson: Gson
+) {
+
+  fun update(crn: String, calculationId: UUID) {
+    val tierChangeEvent = TierChangeEvent(EventType.HMPPS_TIER_CALCULATION_COMPLETE, crn, calculationId)
+    val message = Message(gson.toJson(tierChangeEvent), calculationId.toString())
```
markreesmoj, line 18: It looks like we are toJsoning tierChangeEvent twice - once on this line and then on the next. Do we need to?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/MandateForChangeTest.kt
```diff
@@ -20,11 +27,11 @@ class MandateForChangeTest : MockedEndpointsTestBase() {
...
     setupMaleOffenderWithRegistrations(crn)
 
-    val expectedTierUpdate = tierUpdateWillSucceed(crn, "A0")
-
     listener.listen(calculationMessage(crn))
+    val tier = repo.findFirstByCrnOrderByCreatedDesc(crn)
 
-    mockCommunityApiServer.verify(expectedTierUpdate)
+    Assertions.assertThat(tier?.data?.change?.tier).isEqualTo(ChangeLevel.ZERO)
+    Assertions.assertThat(tier?.data?.protect?.tier).isEqualTo(ProtectLevel.A)
```
markreesmoj, line 34: This would test more of our code if it checked the write to the SNS topic as well. I'm not sure how easy that would be though

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/NoSentenceTest.kt
```diff
@@ -21,7 +21,6 @@ class NoSentenceTest : MockedEndpointsTestBase() {
     setUpNoSentence(crn)
     setupRegistrations(registrationsResponse(), crn)
     restOfSetupWithMaleOffenderNoSevereNeeds(crn)
-    tierUpdateWillSucceed(crn, "A0")
```
markreesmoj, line 24: this has an assertion in it, I don't think this test has any assertions any more

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/MandateForChangeTest.kt
markreesmoj: Having read the rest of it, it looks like we are going to lose all of the end to end tests in this PR.

I mean the tests covering the flow from the listener to the SNS topic. I think it would be handy if we had something to give us confidence that all the bits join up properly. 

Maybe a smoke test which we run post-deployment to change a known CRN might be a better option than keeping the end to end tests here? I  do worry that with so much mocking, we can end up binding ourselves quite closely to the implementation without proving that it actually works as expected.

helm_deploy/values-dev.yaml
nickmcmahon01: This project should always emit the event, It's up the clients to decide how to handle events they subscribe to I think. I have raised https://dsdmoj.atlassian.net/browse/MAS-374 to move the feature toggle to the delius-updates service .

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/AwsConfiguration.kt
nickmcmahon01: I'll double check this from the secrets down to the property

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/SuccessUpdater.kt
nickmcmahon01: Good spot, i've copied it the domain from the client project and there is it ? because its an enum of either 'ones we care about or null' if it was a String it would always be populated in the client too.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/SuccessUpdater.kt
nickmcmahon01: I _think_ so because the message is a json object and the parcel is also a json object with a string message.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/MandateForChangeTest.kt
nickmcmahon01: Shal we bite the bullet and move away from mocked sns/sqs and run it on localstack on the build?

helm_deploy/values-dev.yaml
markreesmoj: The main thing is that we don't accidentally enable writeback when we deploy this to production

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/MandateForChangeTest.kt
markreesmoj: I think we've a good chance of succeeding with that. It was easy on the sqs tool

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/NoSentenceTest.kt
nickmcmahon01: good spot, I thought I did them all, I've added the same repo assertion as the rest for now.

markreesmoj (approved):
👍

nickmcmahon01:
Agreed to split out the test tech debt work into
https://dsdmoj.atlassian.net/browse/MAS-375.

PR #73, opened 2021-03-24
Fix the message format for sending to the SNS

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/SuccessUpdater.kt
```diff
@@ -14,19 +14,22 @@ class SuccessUpdater(
 ) {
 
   fun update(crn: String, calculationId: UUID) {
-    val tierChangeEvent = TierChangeEvent(EventType.TIER_CALCULATION_COMPLETE, crn, calculationId)
-    val message = Message(gson.toJson(tierChangeEvent), calculationId.toString())
+    val message = Message(EventType.TIER_CALCULATION_COMPLETE, gson.toJson(TierChangeEvent(crn, calculationId)), calculationId.toString())
+
```
markreesmoj, line 18: maybe log the message here just in case

markreesmoj (approved):

PR #74, opened 2021-03-26
Filter explicitly for assessments with COMPLETE status instead of jus…

nickmcmahon01:
…t relying on a completed date.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
@@ -6,12 +6,12 @@ import io.swagger.annotations.ApiResponse
 import io.swagger.annotations.ApiResponses
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
 import org.springframework.http.ResponseEntity
-import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.hmppstier.dto.TierDto
+import uk.gov.justice.digital.hmpps.hmppstier.service.SuccessUpdater
 import uk.gov.justice.digital.hmpps.hmppstier.service.TierCalculationService
 import uk.gov.justice.digital.hmpps.hmppstier.service.exception.EntityNotFoundException
 import java.util.UUID
@@ -19,8 +19,7 @@ import java.util.UUID
 @Api
 @RestController
 @RequestMapping(produces = [APPLICATION_JSON_VALUE])
-@PreAuthorize("hasRole('ROLE_HMPPS_TIER')")
-class TierCalculationController(private val tierCalculationService: TierCalculationService) {
+class TierCalculationController(private val tierCalculationService: TierCalculationService, private val successUpdater: SuccessUpdater) {
 
   @ApiOperation(value = "Retrieve tiering score by crn")
   @ApiResponses(
@@ -31,8 +30,9 @@ class TierCalculationController(private val tierCalculationService: TierCalculat
   )
 
   @GetMapping("crn/{crn}/tier")
-  fun getLatestTierCalculation(@PathVariable(required = true) crn: String): ResponseEntity<TierDto> {
-    return ResponseEntity.ok(tierCalculationService.getLatestTierByCrn(crn))
+  fun getLatestTierCalculation(@PathVariable(required = true) crn: String): ResponseEntity<String> {
+    successUpdater.update(crn, UUID.randomUUID())
+    return ResponseEntity.ok("ok")
       ?: throw EntityNotFoundException("Tier Result Not Found for $crn")
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiService.kt
@@ -19,7 +19,8 @@ class AssessmentApiService(
   fun getRecentAssessment(crn: String): OffenderAssessment? =
     assessmentApiClient.getAssessmentSummaries(crn)
       .filter {
-        it.voided == null &&
+        it.assessmentStatus == "COMPLETE" &&
+          it.voided == null &&
           it.completed != null &&
           it.completed.toLocalDate().isAfter(LocalDate.now(clock).minusWeeks(55).minusDays(1))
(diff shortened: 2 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiService.kt
```diff
@@ -19,7 +19,8 @@ class AssessmentApiService(
   fun getRecentAssessment(crn: String): OffenderAssessment? =
     assessmentApiClient.getAssessmentSummaries(crn)
       .filter {
-        it.voided == null &&
+        it.assessmentStatus == "COMPLETE" &&
+          it.voided == null &&
           it.completed != null &&
```
markreesmoj, line 24: Do we still need thic check if assessmentStatus == "COMPLETE"?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -31,8 +30,9 @@ class TierCalculationController(private val tierCalculationService: TierCalculat
   )
 
   @GetMapping("crn/{crn}/tier")
-  fun getLatestTierCalculation(@PathVariable(required = true) crn: String): ResponseEntity<TierDto> {
-    return ResponseEntity.ok(tierCalculationService.getLatestTierByCrn(crn))
+  fun getLatestTierCalculation(@PathVariable(required = true) crn: String): ResponseEntity<String> {
```
markreesmoj, line 33: This doesn't look related to the rest of the PR, but if it is supposed to be here then we could do with a test for it

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculatorTest.kt
```diff
@@ -60,7 +60,7 @@ internal class ChangeLevelCalculatorTest {
 
     @Test
     fun `should calculate Oasys Needs none`() {
-      val assessment = OffenderAssessment("12345", LocalDateTime.now(clock), null)
+      val assessment = OffenderAssessment("12345", LocalDateTime.now(clock), null, null)
```
markreesmoj, line 63: This looks a bit misleading - if the status is null then the assessment needs would not be retrieved as we would have ditched the assessments by this stage I think? I know it doesn't affect the test.

markreesmoj (changes requested):
The assessments-api call change looks OK - I didn't understand the change to the TierCalculationController though

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiService.kt
nickmcmahon01: I'm not sure, for me its belts and braces, we need to check it for null on the 55 week check anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
nickmcmahon01: I accidently checked this in. I'll revert.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculatorTest.kt
nickmcmahon01: I'm happy to change it, for me in unit tests I use `null` to say "I don't care about this value here" the filtering is done in another part of the code and this code just needs to accept whatever is passed in, we could use `"AnyStatus"` instead of null?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AssessmentApiService.kt
nickmcmahon01: I'll combine the checks.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -6,21 +6,20 @@ import io.swagger.annotations.ApiResponse
...
 import uk.gov.justice.digital.hmpps.hmppstier.service.TierCalculationService
 import uk.gov.justice.digital.hmpps.hmppstier.service.exception.EntityNotFoundException
 import java.util.UUID
 
 @Api
 @RestController
 @RequestMapping(produces = [APPLICATION_JSON_VALUE])
-@PreAuthorize("hasRole('ROLE_HMPPS_TIER')")
-class TierCalculationController(private val tierCalculationService: TierCalculationService) {
+class TierCalculationController(private val tierCalculationService: TierCalculationService, private val successUpdater: SuccessUpdater) {
```
markreesmoj, line 22: This class shouldn't even exist, should it? It's been moved to the delius-update project

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -6,21 +6,20 @@ import io.swagger.annotations.ApiResponse
 import io.swagger.annotations.ApiResponses
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
 import org.springframework.http.ResponseEntity
-import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.hmppstier.dto.TierDto
+import uk.gov.justice.digital.hmpps.hmppstier.service.SuccessUpdater
```
markreesmoj, line 14: I don't think this should be here

markreesmoj (commented):
I still think there is some code in the controller which doesn't need to be ther

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
nickmcmahon01: I'd like that functionality to be distinct from the tier calculation functionality, what should we wall the class? tierToDeliusUpdatesClient ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
nickmcmahon01: my mistake.

PR #82, opened 2021-04-01
Store a breakdown of points along with the total for 3rd line support

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -95,6 +98,7 @@ class ProtectLevelCalculator(
       .distinct()
       .filter { it != ComplexityFactor.IOM_NOMINAL }
       .count()
+      .times(2)
```
markreesmoj, line 101: Is this a new requirement?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -104,7 +108,8 @@ class ProtectLevelCalculator(
   ): Int =
     when {
       communityApiClient.getOffender(crn).gender.equals("female", true) -> {
-        getAdditionalFactorsAssessmentComplexityPoints(offenderAssessment) + getBreachRecallComplexityPoints(crn, convictions)
+        (getAdditionalFactorsAssessmentComplexityPoints(offenderAssessment) + getBreachRecallComplexityPoints(crn, convictions))
+          .times(2)
```
markreesmoj, line 112: Is this a new requirement?

src/main/resources/application.yml
```diff
@@ -74,3 +74,6 @@ management:
     info:
       cache:
         time-to-live: 2000ms
+
+calculation:
+  version:  1
```
markreesmoj, line 79: We should probably put a note in the README about when to change this, whether we are using semver or whole numbers etc

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
nickmcmahon01: I've moved them from calling method into the callee as they're two separate requirements that we store the results of individually.

markreesmoj: 👍

markreesmoj (approved):

PR #86, opened 2021-04-06
MAS-384 Finishing test refactor

markreesmoj:
remove wiremock
more convenience methods for integration tests


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/setup/MockedEndpointsTestBase.kt
```diff
@@ -239,12 +208,12 @@ abstract class MockedEndpointsTestBase : IntegrationTestBase() {
   }
 }
 
-data class TierChangeEvent(
+private data class TierChangeEvent(
```
nickmcmahon01, line 211: Is it worth using the real ones?

markreesmoj: Excellent idea

nickmcmahon01 (approved):

PR #88, opened 2021-04-06
MAS-391 Ignore additional requirements

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationRequiredEventListener.kt
```diff
@@ -14,6 +15,12 @@ class TierCalculationRequiredEventListener(
   private val calculator: TierCalculationService
 ) {
 
+  @MessageExceptionHandler()
```
markreesmoj, line 18: This will hopefully help us investigate why calculations have failed

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/NeedsTest.kt
```diff
@@ -11,7 +11,7 @@ class NeedsTest : MockedEndpointsTestBase() {
 
   @Test
   fun `severe needs 18 points plus 2 OGRS make change level 3`() {
-    val crn = "X333444"
+    val crn = "X333445"
```
markreesmoj, line 14: This is a bit dodgy - if we reuse CRNs across integration tests then one of the tests fails. Should look into why that is really

src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markreesmoj, line 1: copy/paste of the main one with a few extra lines. I tried to import the main one using include resource but didn't have much luck

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationRequiredEventListener.kt
nickmcmahon01: noice

nickmcmahon01 (approved):

PR #89, opened 2021-04-06
MAS-391 ignore inactive requirements

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -99,7 +98,7 @@ class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val we
       .uri("/offenders/crn/$crn")
       .retrieve()
       .bodyToMono(Offender::class.java)
-      .block() ?: throw EntityNotFoundException("No Offender record found for $crn")
+      .block()
```
markreesmoj, line 101: I think this will either return an offender, 404 or 500. In the case of a 404 it now logs

2021-04-07 08:56:46.249  INFO 32803 --- [enerContainer-2] lculationRequiredEventListener$Companion : Failed to calculate tier for CRN X123456 with error: 404 Not Found from GET http://localhost:8091/secure/offenders/crn/X123456/assessments |

markreesmoj: Hmm, that is the assessments endpoint. Well the same logic applies I think!

nickmcmahon01 (approved):

PR #91, opened 2021-04-08
Assessment is always present by this point

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculator.kt
```diff
@@ -39,7 +39,7 @@ class ChangeLevelCalculator(
 
       else -> {
         val points = mapOf(
-          NEEDS to getAssessmentNeedsPoints(offenderAssessment),
+          NEEDS to getAssessmentNeedsPoints(offenderAssessment!!.assessmentId),
```
nickmcmahon01, line 42: can we take this a step back and remove the ? from the client DTO?
It feels like we should always have a assessmentId but you'd be amazed.

markreesmoj: I'll try to push it up through the code. It can be null when there's no assessments but maybe the nullobject pattern could help? I'll see what happens.

nickmcmahon01 (approved):

PR #93, opened 2021-04-09
MAS-393 requirements are always active

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculatorTest.kt
```diff
@@ -350,33 +350,6 @@ internal class ProtectLevelCalculatorTest {
       validate()
     }
 
-    @Test
```
markreesmoj, line 353: RLRH is not in the ROSH enum so was always ignored anyway

nickmcmahon01 (approved):

PR #95, opened 2021-04-15
MAS-398 do not override oauth scopes

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/setup/MockedEndpointsTestBase.kt
```diff
@@ -72,15 +70,29 @@ abstract class MockedEndpointsTestBase : IntegrationTestBase() {
     oauthMock.stop()
   }
 
+  private fun setupOauth() {
+    val response = HttpResponse.response().withContentType(APPLICATION_JSON)
+      .withBody(gson.toJson(mapOf("access_token" to "ABCDE", "token_type" to "bearer")))
+    oauthMock.`when`(request().withPath("/auth/oauth/token").withBody("grant_type=client_credentials")).respond(response)
```
markreesmoj, line 76: the exact match on the body here will fail if scopes are reintroduced

andymarke (approved):
LGTM

PR #98, opened 2021-04-19
MAS-390 mandate for change exclude order length extended

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/MandateForChange.kt
```diff
@@ -13,27 +13,29 @@ class MandateForChange(
...
   fun hasNoMandate(crn: String, convictions: Collection<Conviction>): Boolean =
     convictions
-      .filter { it.sentence?.terminationDate == null }
-      .let { activeConvictions ->
-        activeConvictions.none {
-          it.sentence != null && (isCustodial(it.sentence) || hasNonRestrictiveRequirements(crn, it.convictionId))
-        }
+      .filter { currentSentence(it.sentence) }
+      .none {
+        isCustodial(it.sentence) || hasNonRestrictiveRequirements(crn, it.convictionId)
```
nickmcmahon01, line 18: nitpick but this could be 
`.none {
     currentSentence(it.sentence) && ( isCustodial(it.sentence) || hasNonRestrictiveRequirements(crn, it.convictionId))`

Not sure what is more readable.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/MandateForChange.kt
```diff
@@ -13,27 +13,29 @@ class MandateForChange(
...
       }.also { log.debug("Has no mandate for change: $it") }
 
-  private fun isCustodial(sentence: Sentence) =
-    sentence.sentenceType.code in custodialSentences
+  private fun currentSentence(sentence: Sentence?) =
+    sentence != null && sentence.terminationDate == null
+
+  private fun isCustodial(sentence: Sentence?) =
+    sentence?.sentenceType?.code in custodialSentences
```
nickmcmahon01, line 25: nickpick but I wonder if we're being too accommodating in general, We'd know at parse time if we don't have a sentenceType for example, if we need one we can throw straight away and it will go to the DQL.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/MandateForChange.kt
```diff
@@ -13,27 +13,29 @@ class MandateForChange(
...
 
-  private fun excludeUnpaidWork(it: Requirement) =
-    it.requirementTypeMainCategory?.code != "W"
+  private fun excludeUnpaidWork(it: Requirement): Boolean =
+    it.requirementTypeMainCategory?.code !in unpaidWorkAndOrderExtended
 
   companion object {
     private val log = LoggerFactory.getLogger(this::class.java)
     private val custodialSentences = arrayOf("NC", "SC")
+    private val unpaidWorkAndOrderExtended = arrayOf("W", "W1")
```
nickmcmahon01, line 39: Was there a W2?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/MandateForChange.kt
markreesmoj: Definitely agree with this. I did a little PR in between times to try to reduce the amount of null checking we do. It gets rid of the first check here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/MandateForChange.kt
markreesmoj: Not on the offender provided. I have seen no evidence of W2 as yet

nickmcmahon01 (approved):

PR #99, opened 2021-04-19
We only want current convictions so we can filter them when retrieving

markreesmoj:
This reduces complexity and duplication (two null checks for sentence down to one). 
It does introduce another object though (ConvictionDto and Conviction).
We could get rid of quite a bit of complexity if we did more of this kind of filtering

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -98,11 +98,11 @@ tasks {
       rule {
         limit {
           counter = "BRANCH"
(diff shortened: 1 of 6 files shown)


markreesmoj:
I will have a look tomorrow morning for any similar stuff so we can go through them with Ed

nickmcmahon01 (approved):

PR #101, opened 2021-04-21
Requirement type main category

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/AdditionalRequirementsTest.kt
```diff
@@ -11,12 +11,12 @@ class AdditionalRequirementsTest : MockedEndpointsTestBase() {
...
   fun `Additional requirements do not cause a processing error`() {
-    val crn = "X333444"
+    val crn = "X833444"
     setupNonCustodialSentence(crn)
     setupRegistrations(registrationsResponse(), crn)
     setupAdditionalRequirements(crn)
     restOfSetupWithMaleOffenderAndSevereNeeds(crn)
     calculateTierFor(crn)
-    expectTierCalculation("A3")
+    expectTierCalculation("A0")
```
markreesmoj, line 20: Surprise!

I wrote this test some time ago and all I wanted was for it to calculate a tier, as this previously threw an error. I should have spent a bit more time looking into what the tier actually was.

I think there's a case for ditching a lot of the existing integration tests and replacing them with BDD. There may be more incorrect behaviour like this which is reinforced by testing.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -54,6 +54,8 @@ class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val we
 
   fun getRequirements(crn: String, convictionId: Long): List<Requirement> {
     return getRequirementsCall(crn, convictionId)
+      .filterNot { it.requirementTypeMainCategory == null && it.restrictive == null }
+      .map { Requirement(it.restrictive!!, it.requirementTypeMainCategory!!.code) }
```
nickmcmahon01, line 58: It looks like our service layer is emerging in the client, and we're creating a domain separate from the DTOs

nickmcmahon01 (approved):

PR #105, opened 2021-04-21
Nm arson violence

nickmcmahon01:


markreesmoj (approved):
👍

helm_deploy/hmpps-tier/templates/_envs.tpl
```diff
@@ -34,6 +34,15 @@ env:
   - name: ASSESSMENT_ENDPOINT_URL
     value: "{{ .Values.env.ASSESSMENT_ENDPOINT_URL }}"
 
+  - name: CALCULATION_VERSION
```
markreesmoj, line 37: Decoupling the version from these two flags seems a bit risky. We could potentially enable one of the flags and not change the version number, which would potentially invalidate calculations. It certainly introduces more complexity.

I wonder if we could just use the version number for the necessary conditional? Is there anything to say that we can't do Arson and Violence as 1.1 and Ten Months Plus as 1.2 for example

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -104,8 +108,20 @@ class ProtectLevelCalculator(
   ): Int =
     when {
       isFemale(crn) -> {
-        (getAdditionalFactorsAssessmentComplexityPoints(offenderAssessment) + getBreachRecallComplexityPoints(crn, convictions))
-          .times(2)
+        val additionalFactorsPoints = getAdditionalFactorsAssessmentComplexityPoints(offenderAssessment)
+        val breachRecallPoints = getBreachRecallComplexityPoints(crn, convictions)
+
+        val violenceArson = when {
```
markreesmoj, line 114: Small matters now but half of these values are suffixes with 'Points' and half are not. We should choose one, don't mind which.

Also half of the calculations are abstracted away in functions and half are not. Again we should choose one, and here I have a preference for abstraction on the grounds of readability.

An alternative might even be to wrap the three calculations based on the convictions into a single method? It might be difficult to name though

The root cause is that there is a great deal of conplexity in this class

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculatorTest.kt
```diff
@@ -1056,4 +1061,31 @@ internal class ProtectLevelCalculatorTest {
       verify { communityApiClient.getOffender(crn) }
     }
   }
+
+  @Nested
+  @DisplayName("Arson and Violence tests")
+  inner class ArsonViolenceTests {
+    @Test
+    fun `Should respect Arson & Violence Toggle`() {
```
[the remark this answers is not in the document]
nickmcmahon01: It's not a great test, but they're set up in this class when we create the service they're passed in as false,false

helm_deploy/hmpps-tier/templates/_envs.tpl
nickmcmahon01: I considered this and thought it was too complex having to do `$version >= 1.1` checks and such, but thinking about it again, we'll just remove the feature flag once we're happy so we won't litter the codebase with complex checks.

PR #106, opened 2021-04-21
BDD tests

markreesmoj:


markreesmoj:
This is in a right mess. At the moment, I cannot run the cucumber test

This article is helpful
https://ultraclearance.com/spring-boot-kotlin-guide-cucumber-black-box-tests-with-external-services-simulated-by-use-of-wiremock/
but it uses maven and junit4 vs gradle and junit5

This one shows cucumber and junit5 with gradle
https://blog.cronn.de/en/testing/2020/08/17/cucumber-junit5.html

And this one cucumber and kotlin
https://medium.com/@mlvandijk/kukumber-getting-started-with-cucumber-in-kotlin-e55112e7309b

Of all the possible culprits, I am inclined to think that gradle is the most likely

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculatorTest.kt
```diff
@@ -71,24 +71,6 @@ internal class ProtectLevelCalculatorTest {
   @DisplayName("Simple Risk tests")
   inner class SimpleRiskTests {
 
-    @Test // RsrThresholds.TIER_B_RSR
```
markreesmoj, line 74: There are BDD tests for this now

nickmcmahon01:
This looks great, I get its a bit of a nonsense adding this in after the fact but we can use it for future work now that it's plumbed in.

nickmcmahon01 (approved):

PR #111, opened 2021-04-29
Add missing RSR test

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -61,8 +61,8 @@ class ProtectLevelCalculator(
     deliusAssessments?.rsr
       .let { rsr ->
         when {
-          rsr != null && rsr >= RsrThresholds.TIER_B_RSR.num -> 20
-          rsr != null && rsr >= RsrThresholds.TIER_C_RSR.num -> 10
+          rsr != null && rsr != RsrThresholds.NO_RSR_MAGIC_NUMBER.num && rsr >= RsrThresholds.TIER_B_RSR.num -> 20
```
markreesmoj, line 64: I thought we had done this too!
Is there a nice kotlin way we can reduce the duplication here?

markreesmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
nickmcmahon01: why yes, there is.

PR #117, opened 2021-05-04
Add Ogrs feature

nickmcmahon01:


src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/OGRS.feature
```diff
@@ -0,0 +1,72 @@
+#lang: en
+Feature: OGRS
+
+  Scenario: Take 10's less than 10%
```
markreesmoj, line 4: I don't understand the scenario title here. If it means something to the business that's not a problem though

src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/criminogenicNeeds.feature
```diff
@@ -0,0 +1,170 @@
+#lang: en
+Feature: Criminogenic Needs
+
+Given the following assessment needs:
```
markreesmoj, line 4: I am pretty sure that these tests don't run in the build because they don't have a Scenario

markreesmoj (changes requested):
Just need to make sure that the needs tests run

src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/OGRS.feature
nickmcmahon01: Its from Stephan, I  can rephrase to "Ogrs score of xxx scores x points"?

src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/criminogenicNeeds.feature
nickmcmahon01: I think I agree since the second to last one has a mistake on it

markreesmoj (approved):
👍

PR #125, opened 2021-05-10
Remove duplicate Mandate For Change tests

markreesmoj:
Also use int arguments where possible

### DIFF
--- a/src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/mandateForChange.9.5.feature
+++ b/src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/mandateForChange.9.5.feature
@@ -6,13 +6,13 @@ Feature: Mandate for change
     Given an offender with a current sentence of type 'SC'
     And an OGRS score of "50"%
     When a tier is calculated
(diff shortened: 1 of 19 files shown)


src/test/resources/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/mandateForChange.9.5.feature
```diff
@@ -6,13 +6,13 @@ Feature: Mandate for change
     Given an offender with a current sentence of type 'SC'
     And an OGRS score of "50"%
     When a tier is calculated
-    Then there is a mandate for change and a change level of "1" is returned for "5" points
+    Then there is a mandate for change and a change level of 1 is returned and 5 points are scored
```
nickmcmahon01, line 9: BDD picky, but should we not just be asserting <> level 1 here  as 'There is mandate for change'

nickmcmahon01:
```
Given an offender with a current sentence of type 'SC'
When a tier is calculated
Then there is a mandate for change
```
'Then there is mandate for change: changeLevel != 0

PR #128, opened 2021-05-13
Simplifying registrations

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -7,18 +7,17 @@ import org.springframework.beans.factory.annotation.Qualifier
...
 
-  fun getRegistrations(crn: String): Collection<Registration> {
-    return getRegistrationsCall(crn).sortedByDescending { it.startDate }
-      .also {
-        log.info("Fetched ${it.size} Registrations for $crn")
-      }
-  }
+  fun getRegistrations(crn: String): Pair<List<Registration>, List<Registration>> =
+    getRegistrationsCall(crn).sortedByDescending { it.startDate }
+      .partition { ComplexityFactor.from(it.type.code) == IOM_NOMINAL }
```
markreesmoj, line 20: Doing this once rather than twice

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ChangeLevelCalculator.kt
```diff
@@ -62,8 +60,8 @@ class ChangeLevelCalculator(
     (
       assessmentApiService.getAssessmentNeeds(assessmentId)
         .let {
-          it.entries.sumBy { ent ->
-            ent.key.weighting.times(ent.value?.score ?: 0)
+          it.entries.sumOf { ent ->
```
nickmcmahon01, line 63: noice

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
nickmcmahon01: I can see the benefit but I think this worsens coupling and cohesion, having a client that knows about change and protect

nickmcmahon01 (approved):

PR #135, opened 2021-05-14
Introduce a Delius Assessments Domain object

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -15,16 +14,15 @@ import java.time.LocalDate
 class CommunityApiClient(@Qualifier("communityWebClientAppScope") private val webClient: WebClient) {
 
   fun getRegistrations(crn: String): Pair<List<Registration>, List<Registration>> =
-    getRegistrationsCall(crn).sortedByDescending { it.startDate }
-      .partition { ComplexityFactor.from(it.type.code) == IOM_NOMINAL }
+    getRegistrationsCall(crn).partition { it.type.code == IOM_NOMINAL.registerCode }
```
nickmcmahon01, line 17: The SortedBy is a nonsense that we put in before we understood the requirements.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -15,16 +14,15 @@ import java.time.LocalDate
...
 
-  fun getDeliusAssessments(crn: String): DeliusAssessments? {
+  fun getDeliusAssessments(crn: String): DeliusAssessments {
     return webClient
       .get()
       .uri("/offenders/crn/$crn/assessments")
       .retrieve()
       .bodyToMono(DeliusAssessments::class.java)
-      .block()
+      .block() ?: DeliusAssessments(BigDecimal(0), 0)
```
nickmcmahon01, line 25: This is what I don't like, but it tidies up everything downstream of it

nickmcmahon01: I don't think its the right solution, but if we translated this DTO into a domain object with sensible defaults we'd get the same result.

markreesmoj: I've been thinking about doing this for a while

markreesmoj:
I like the direction this is going in. I think the reason why so much is failing is that neither rsr nor OGRS is guaranteed to be present on the response from community-api. So handling that by defaulting to 0 and 0.0 should sort this out

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/SetupData.kt
```diff
@@ -63,9 +63,9 @@ class SetupData(
   private var rosh: String = "NO_ROSH"
   private var rsr: BigDecimal = BigDecimal(0)
   private var assessmentAnswers: MutableMap<String, String> = mutableMapOf(
-    Pair(IMPULSIVITY.answerCode, "0"),
-    Pair(TEMPER_CONTROL.answerCode, "0"),
-    Pair(PARENTING_RESPONSIBILITIES.answerCode, "NO")
+    IMPULSIVITY.answerCode to "0",
+    TEMPER_CONTROL.answerCode to "0",
+    PARENTING_RESPONSIBILITIES.answerCode to "NO"
```
markreesmoj, line 68: kotlin power

markreesmoj (approved):
Nice work. I'm not merging it just yet as we haven't released to production for a little while and I would like to clear the way first

PR #147, opened 2021-05-21
Take any 'application logic' out of ProtectTierCalculator

nickmcmahon01:


build.gradle.kts
```diff
@@ -129,7 +129,7 @@ tasks {
       rule {
         limit {
           counter = "BRANCH"
-          minimum = BigDecimal(0.89)
+          minimum = BigDecimal(0.88)
```
nickmcmahon01, line 132: I did get rid of some BS tests..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -37,17 +36,19 @@ class ProtectLevelCalculator(
 
   fun calculateProtectLevel(
     crn: String,
-    offenderAssessment: OffenderAssessment?,
-    rsrScore: BigDecimal,
-    registrations: Collection<Registration>,
-    convictions: Collection<Conviction>
+    offenderAssessment: OffenderAssessment? = null,
+    rsrScore: BigDecimal = BigDecimal.ZERO,
```
markreesmoj, line 40: rsrScore is already defaulted here:
https://github.com/ministryofjustice/hmpps-tier/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/domain/DeliusAssessments.kt#L14
We shouldn't default it again here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -37,17 +36,19 @@ class ProtectLevelCalculator(
...
+    rosh: Rosh? = null,
+    mappa: Mappa? = null,
+    complexityFactors: Collection<ComplexityFactor> = listOf(),
+    convictions: Collection<Conviction> = listOf()
```
markreesmoj, line 44: This I think is also always going to be an empty list so should not be defaulted

https://github.com/ministryofjustice/hmpps-tier/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/CommunityApiService.kt#L17

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -69,35 +70,21 @@ class ProtectLevelCalculator(
...
-          M1 -> 5
-          else -> 0
-        }
-      }
+  private fun getMappaPoints(mappa: Mappa?): Int =
+    when (mappa) {
+      M3, M2 -> 30
+      M1 -> 5
+      else -> 0
+    }
```
markreesmoj, line 86: Much nicer in here now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -37,17 +36,19 @@ class ProtectLevelCalculator(
...
     crn: String,
-    offenderAssessment: OffenderAssessment?,
-    rsrScore: BigDecimal,
-    registrations: Collection<Registration>,
-    convictions: Collection<Conviction>
+    offenderAssessment: OffenderAssessment? = null,
+    rsrScore: BigDecimal = BigDecimal.ZERO,
+    rosh: Rosh? = null,
+    mappa: Mappa? = null,
+    complexityFactors: Collection<ComplexityFactor> = listOf(),
```
markreesmoj, line 43: I think this will always be present too

build.gradle.kts
markreesmoj: I am very happy to see the collaborators test go. Even so this is the second time this value has gone down, and whilst last time it was probably a rounding error this time this time there are three uncovered branches
 
 https://1965-327621961-gh.circle-artifacts.com/0/coverage/uk.gov.justice.digital.hmpps.hmppstier.service/index.html
 
 vs
 
 https://1965-327621961-gh.circle-artifacts.com/0/coverage/uk.gov.justice.digital.hmpps.hmppstier.service/index.html
 
 Specifically the newly uncovered branches are here, somewhat to my surprise 
 https://1967-327621961-gh.circle-artifacts.com/0/coverage/uk.gov.justice.digital.hmpps.hmppstier.service/TierCalculationService.kt.html#L44

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/CommunityApiService.kt
```diff
@@ -15,4 +19,13 @@ class CommunityApiService(
 
   fun getConvictionsWithSentences(crn: String): List<Conviction> =
     communityApiClient.getConvictions(crn).filterNot { it.sentence == null }.map { Conviction.from(it) }
+
```
markreesmoj, line 22: I'm not 100% sure these methods belong on this, although they might.
CommunityApiService was previously an n-tier style service layer on top of the Client, now it is also doing transformations on data retrieved directly from the client.
I know it's a lot of work and definitely not one for this PR but how about a single getRegistrations method which returns these three transformed properties on the object?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/CommunityApiServiceTest.kt
```diff
@@ -0,0 +1,335 @@
...
+      Assertions.assertThat(result).isEqualTo(Rosh.MEDIUM)
+    }
+  }
+
+  @Nested
+  @DisplayName("Simple Mappa tests")
+  inner class SimpleMappaTests {
+
+    @Test
+    fun `should return null for No Rosh`() {
```
markreesmoj, line 137: Rosh should be Mappa

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationServiceTest.kt
```diff
@@ -143,75 +140,4 @@ internal class TierCalculationServiceTest {
...
-      every { changeLevelCalculator.calculateChangeLevel(crn, any(), any(), any(), any(), any()) } returns changeLevelResult
-
-      every { tierCalculationRepository.findFirstByCrnOrderByCreatedDesc(crn) } returns validTierCalculationEntity
-
-      val slot = slot<TierCalculationEntity>()
-      every { tierCalculationRepository.save(capture(slot)) } answers { slot.captured }
-
-      service.calculateTierForCrn(crn)
-
-      // We don't update the SNS and recognise that is hasn't change.
```
markreesmoj, line 169: I think this is where the coverage has dropped.
I am not keen on seeing these tests come back though. Maybe there's another way using integration tests or BDD - causing two tier calculations perhaps?

markreesmoj (changes requested):
Sorry it took me so long - this PR is quite large.

I think the default values in the ProtectLevelCalculator are a bit misleading and would like to see if we can do without them. The Mappa and Rosh defaults are probably OK for now

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationServiceTest.kt
nickmcmahon01: There is some BDD missing around "And an Update event is generated"

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/CommunityApiServiceTest.kt
nickmcmahon01: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
nickmcmahon01: I agree, its finding the right balance between these and having loads of nonsense in the tests. I wonder if Kotlin can help.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/CommunityApiService.kt
nickmcmahon01: Yes it could be a domain object.

build.gradle.kts
nickmcmahon01: I'll look at these.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelTwoCalculatorTest.kt
```diff
@@ -47,6 +50,10 @@ internal class ProtectLevelTwoCalculatorTest {
 
   private val crn = "Any Crn"
 
+  private fun calculateProtectLevel(crn: String, offenderAssessment: OffenderAssessment? = null, rsrScore: BigDecimal = BigDecimal.ZERO, rosh: Rosh? = null, convictions: Collection<Conviction> = listOf()): TierLevel<ProtectLevel> {
+    return service.calculateProtectLevel(crn, offenderAssessment, rsrScore, rosh, null, listOf(), convictions)
+  }
```
markreesmoj, line 55: 🥇

markreesmoj (approved):
👍

PR #152, opened 2021-05-25
Mas 422 l3

markreesmoj:
Work in progress
It is tricky to run two different versions of the application

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -101,8 +101,13 @@ tasks {
       .singleFile
     jvmArgs = listOf("-javaagent:$jacocoAgent=destfile=$buildDir/jacoco/cucumber.exec,append=false")
   }
+  val versionTwo by registering(Test::class) {
+    include("**/ProtectTierATest*")
+    finalizedBy(cucumber)
+  }
   getByName("check") {
-    dependsOn(":ktlintCheck")
+    dependsOn(":ktlintCheck", detekt)
+    finalizedBy(versionTwo)
   }
   getByName<JacocoReport>("jacocoTestReport") {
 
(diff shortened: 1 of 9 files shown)


build.gradle.kts
```diff
@@ -101,8 +101,13 @@ tasks {
       .singleFile
     jvmArgs = listOf("-javaagent:$jacocoAgent=destfile=$buildDir/jacoco/cucumber.exec,append=false")
   }
+  val versionTwo by registering(Test::class) {
```
nickmcmahon01, line 104: This is getting messy, lets sort this out Thursday.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -72,19 +74,21 @@ class ProtectLevelCalculator(
...
 
   private fun getMappaPoints(mappa: Mappa?): Int =
     when (mappa) {
-      M3, M2 -> 30
+      M3, M2 -> levelA()
       M1 -> 5
       else -> 0
     }
 
+  private fun levelA() = if (calculationVersion == 2) 300 else 30
```
nickmcmahon01, line 90: I might be picking on the name, but the points and the level are different things.
What's the 300 for? it is a magic number?

markreesmoj: yes, 30 and 300 are magic numbers. We will see what happens today but this is a way of keeping the link between points scored and protect tier

build.gradle.kts
markreesmoj: It's a nightmare. We also need to consider the BDD tests which are an accident waiting to happen. We have got three versions of the application now and the tests cannot keep up as they are currently structured

markreesmoj:
Merging so we can see what it does in preproduction

PR #153, opened 2021-05-26
Test dynamic version setting

nickmcmahon01:


markreesmoj:
I think we should remove all the code which is not going live any time soon (sentence length and arson/violence), and try out this helper approach for the next time we are asked to try out a calculation just to see what happens. It looks really good for the tests, much better than what's in place for testing at the moment

PR #154, opened 2021-05-26
alternate business rule for Tier fix

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/ProtectLevelCalculator.kt
```diff
@@ -56,14 +57,27 @@ class ProtectLevelCalculator(
...
 
-    return when {
-      total >= levelAThreshold() -> TierLevel(A, total, points)
-      total in 20 until levelAThreshold() -> TierLevel(B, total, points)
-      total in 10 until 20 -> TierLevel(C, total, points)
-      else -> TierLevel(D, total, points)
-    }
+    return TierLevel(
+      when {
+        total >= 30 && hasHighMappaOrVeryHighRosh(points) -> A
```
nickmcmahon01, line 62: I like the red one better IF it goes back to policy and they give us a value, and it's their job to keep the sum of everything else below the value for the new threshold so that the maths always works out ok.
the green one is truer to the current business rule otherwise.

PR #155, opened 2021-05-27
Only write back if the protect or change level has changed

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -42,7 +42,7 @@ class TierCalculationService(
   @Transactional
   fun calculateTierForCrn(crn: String) =
     calculateTier(crn).let {
-      val isUpdated = it.data != getLatestTierCalculation(crn)?.data
+      val isUpdated = isUpdated(it, crn)
```
nickmcmahon01, line 45: this was fine before we added the breakdown of individual scores which has changed the equals().

nickmcmahon01 (approved):

PR #161, opened 2021-06-01
Move additional factors for women up a level

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/TierCalculationService.kt
```diff
@@ -17,14 +17,16 @@ class TierCalculationService(
...
   private val assessmentApiService: AssessmentApiService,
   private val communityApiService: CommunityApiService,
   private val communityApiClient: CommunityApiClient, // Deprecated
   private val successUpdater: SuccessUpdater,
-  private val telemetryService: TelemetryService
+  private val telemetryService: TelemetryService,
+  private val additionalFactorsForWomen: AdditionalFactorsForWomen
 ) {
 
+  private val protectLevelCalculator: ProtectLevelCalculator = ProtectLevelCalculator()
```
nickmcmahon01, line 28: Why treat this one differently to the others? I know we don't do proper IoC, but moving Additional Factors for Women 'up' moves to closer to being able to achieve a hexagonal architecture and this moves us further away.

markreesmoj: A couple of reasons why I thought of doing this:
ProtectLevelCalculator no longer has dependencies so we can
It can no longer be mocked, so tests will have to use the real instance

I'm not clear on what this has to do with hexagonal architecture - this is part of the 'domain' rather than the 'adapters'

However we can certainly keep it as an injected component

markreesmoj: The ProtectLevelCalculatorTest is now much smaller than it was (mostly moved into AdditionalFactorsForWomen) as it is now simpler and is covered by higher level tests. We could perhaps even get rid of that test altogether

It would be useful to draw a target architecture on a whiteboard I think

nickmcmahon01 (approved):

PR #164, opened 2021-06-03
Reduce complexity

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/integration/bdd/BddSteps.kt
```diff
@@ -202,7 +202,6 @@ class BddSteps : En {
     }
     Given("an offender scores 51 protect points") {
       setupData.setGender("Female")
-      setupData.setValidAssessment()
```
markreesmoj, line 205: Done in setAssessmentAnswer instead

nickmcmahon01 (approved):

PR #165, opened 2021-06-03
Protect the endpoints with a role

nickmcmahon01:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/ResourceServerConfiguration.kt
```diff
@@ -36,8 +36,7 @@ class ResourceServerConfiguration : WebSecurityConfigurerAdapter() {
             "/swagger-ui.html",
```
markreesmoj, line 36: not sure why and it's not relevant but I get a 403 for this
https://hmpps-tier-dev.hmpps.service.justice.gov.uk/swagger-ui.html

markreesmoj (approved):
Looks good - I will leave it to you to merge once the updater's client has been updated

nickmcmahon01:
I think it's always had it, we literally just forgot to do this bit...

PR #170, opened 2021-06-04
COMPLEXITY: move hasBreachOrRecall onto CommunityApiService

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AdditionalFactorsForWomenTest.kt
```diff
@@ -29,7 +29,12 @@ class AdditionalFactorsForWomenTest {
   private val clock = Clock.fixed(LocalDateTime.of(2020, 1, 1, 0, 0).toInstant(ZoneOffset.UTC), ZoneId.systemDefault())
   private val communityApiClient: CommunityApiClient = mockk(relaxUnitFun = true)
   private val assessmentApiService: AssessmentApiService = mockk(relaxUnitFun = true)
-  private val additionalFactorsForWomen: AdditionalFactorsForWomen = AdditionalFactorsForWomen(clock, communityApiClient, assessmentApiService)
+  private val additionalFactorsForWomen: AdditionalFactorsForWomen = AdditionalFactorsForWomen(
+    clock,
+    communityApiClient,
+    assessmentApiService,
+    CommunityApiService(communityApiClient)
```
markreesmoj, line 36: Lazy I know but soon this will no longer have the client so I can move the method then

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/AdditionalFactorsForWomen.kt
```diff
@@ -60,7 +60,7 @@ class AdditionalFactorsForWomen(
       .filter { qualifyingConvictions(it.sentence) }
```
nickmcmahon01, line 60: For me a value of the service is that there can be a `getBreachedConvictions` method, and this method becomes the business rule only  `if` any() 2 : 0`

nickmcmahon01 (approved):

nickmcmahon01 (approved):

PR #172, opened 2021-06-13
Overwrite scores without writing back to delius.

nickmcmahon01:


markreesmoj (approved):
As a short term measure to be removed immediately after use I think this will do

PR #182, opened 2021-10-14
WFP-646 started to convert helm chart

carlov20:


helm_deploy/hmpps-tier/values.yaml
```diff
@@ -0,0 +1,62 @@
+---
+generic-service:
+  nameOverride: hmpps-tier
+
+  replicaCount: 1
```
markreesmoj, line 5: 2 hmpps-tier instances are needed

helm_deploy/hmpps-tier/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+  image:
+    repository: quay.io/hmpps/hmpps-tier
+    tag: app_version    # override at deployment time
+    port: 8080
+
+  ingress:
+    enabled: true
+    host: app-hostname.local    # override per environment
+    tlsSecretName: hmpps-tier-cert
+    path: /
```
markreesmoj, line 16: do we need enable-whitelist here?
I don;t really know what I'm doing - just looking for the differences between the old and new versions!

carlov20: its now based off of whether you've defined allowList, no more enable-whitelist required https://github.com/ministryofjustice/hmpps-helm-charts/blob/main/charts/generic-service/templates/ingress.yaml#L14

PR #183, opened 2021-11-09
Wfp 562

carlov20:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/AssessmentApiClient.kt
```diff
@@ -16,10 +16,8 @@ import java.time.LocalDateTime
 @Component
 class AssessmentApiClient(@Qualifier("assessmentWebClientAppScope") private val webClient: WebClient) {
 
-  fun getAssessmentAnswers(assessmentId: String?): Collection<Question> {
-    return assessmentId?.let {
-      getAssessmentAnswersCall(assessmentId)
-    } ?: listOf()
+  fun getAssessmentAnswers(assessmentId: String): Collection<Question> {
```
carlov20, line 19: Upstream call is never null, surely an assessmentId must be supplied?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/advice/ControllerAdvice.kt
```diff
@@ -30,27 +28,13 @@ class ControllerAdvice {
     return ResponseEntity(ErrorResponse(status = 404, developerMessage = e.message), NOT_FOUND)
   }
 
-  @ExceptionHandler(MethodArgumentNotValidException::class)
```
carlov20, line 33: @Valid never used in project. In Javadocs `Exception to be thrown when validation on an argument annotated with @Valid fails.`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/advice/ControllerAdvice.kt
```diff
@@ -30,27 +28,13 @@ class ControllerAdvice {
...
-  }
-
   @ExceptionHandler(HttpMessageConversionException::class)
   @ResponseStatus(BAD_REQUEST)
   fun handle(e: HttpMessageConversionException): ResponseEntity<*> {
     log.error("HttpMessageConversionException: {}", e.message)
     return ResponseEntity(ErrorResponse(status = 400, developerMessage = e.message), BAD_REQUEST)
   }
 
-  @ExceptionHandler(HttpMessageNotReadableException::class)
```
carlov20, line 47: subclass of `HttpMessageConversionException` which is caught above

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/advice/ControllerAdvice.kt
markreesmoj: no doubt copy/paste from another project

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/AssessmentApiClient.kt
markreesmoj: 👍 excellent spot

markreesmoj (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/AssessmentApiClient.kt
trozzomoj: I think assumption  are dangerous ehehhe I would return an empty list in case.

helm_deploy/hmpps-tier/values.yaml
```diff
@@ -22,6 +22,7 @@ generic-service:
     SPRING_PROFILES_ACTIVE: "aws,logstash,stdout"
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    HMPPS_SQS_TOPICS_HMPPSCALCULATIONCOMPLETETOPIC_ASYNC_CLIENT: true
```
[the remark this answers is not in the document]
carlov20: to initialise an AmazonSNSAsync used by the SuccessUpdater already https://github.com/ministryofjustice/hmpps-tier/blob/8f30928c1d01cc67ab3416f42759c738cf178a9c/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/service/SuccessUpdater.kt#L14 .

Docs from SQS library: `If true then the AmazonSNS bean created will be an AmazonSNSAsync instance.`

PR #185, opened 2022-01-19
:wrench: WFP-798: Added ingress to green

katie-cawthorne:


helm_deploy/values-prod.yaml
```diff
@@ -4,6 +4,7 @@
 generic-service:
   ingress:
     host: hmpps-tier.hmpps.service.justice.gov.uk
+    contextColour: green
```
carlov20, line 7: they should all now be green so could you set it https://github.com/ministryofjustice/hmpps-tier/blob/main/helm_deploy/hmpps-tier/values.yaml#L12 block?

carlov20 (approved):
🔥

PR #267, opened 2022-08-25
Reduce dependencies

markreesmoj:
The idea here is to simplify the code, reducing the number of dependencies the TierCalculationService has. 
The outside-in testing we have allowed me to make this change very small, without needing to touch lots of tests

Longer term it would be interesting to see if we might retrieve the data required by these services and pass it in instead of giving them the ability to call out to community and assessment-api.

I am not dead set on merging this, just presenting it as an option for debate - this is a possible way of cutting dependencies by reducing the number of arguments to TierCalculationService (the most complex class in this project)


markreesmoj:
Discussed and approved as a team in slack

PR #306, opened 2022-10-14
PI-463 Only send success event after completed transaction

marcus-bcl:
This PR moves the call to `successUpdater` out of the `@Transactional` service method.  This ensures the TIER_CALCULATION_COMPLETE event is only emitted if the transaction has completed, and also prevents race conditions that occur due to the event being emitted too early (see Slack thread: https://mojdt.slack.com/archives/C035YK9FFK4/p1665733541723359).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationRequiredEventListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationRequiredEventListener.kt
@@ -6,12 +6,14 @@ import org.slf4j.LoggerFactory
 import org.springframework.jms.annotation.JmsListener
 import org.springframework.messaging.handler.annotation.MessageExceptionHandler
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppstier.service.SuccessUpdater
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationRequiredEventListener.kt
```diff
@@ -23,7 +25,9 @@ class TierCalculationRequiredEventListener(
   @JmsListener(destination = "hmppsoffenderqueue", containerFactory = "hmppsQueueContainerFactoryProxy")
   fun listen(msg: String) {
     val (crn) = getCrn(msg)
-    calculator.calculateTierForCrn(crn)
+    calculator.calculateTierForCrn(crn)?.let {
+      successUpdater.update(crn, it.uuid)
+    }
```
markreesmoj, line 30: This looks fine functionally, but it is a bit confusing that a method called `calculateTierForCrn` only returns a calculation when it has changed, even though it always performs and saves that calculation.

sp-watson (approved):

markreesmoj:
Closing - we had a think about this and came up with this smaller change. See https://github.com/ministryofjustice/hmpps-tier/commit/9de2c993e87cb5c481b2fa9d1e3d9cfee95b4603

PR #376, opened 2023-02-28
WFP-1850 use coroutines

carlov20:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
```diff
@@ -2,60 +2,57 @@ package uk.gov.justice.digital.hmpps.hmppstier.client
...
-  fun getConvictions(crn: String): List<ConvictionDto> {
-    val responseType = object : ParameterizedTypeReference<List<ConvictionDto>>() {}
+  suspend fun getConvictions(crn: String): List<ConvictionDto> {
     return webClient
       .get()
       .uri("/offenders/crn/$crn/convictions?activeOnly=true")
       .retrieve()
-      .bodyToMono(responseType)
-      .block() ?: listOf()
+      .bodyToFlow<ConvictionDto>()
```
sp-watson, line 37: I am a bit confused as to why this uses `bodyToFlow` when the one below - which previously used the same `bodyToMono` uses `awaitBody`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/ObjectMapperConfiguration.kt
```diff
@@ -1,27 +0,0 @@
...
-import org.springframework.context.annotation.Bean
-import org.springframework.context.annotation.Configuration
-import org.springframework.context.annotation.Primary
-
-@Configuration
-class ObjectMapperConfiguration {
-
-  @Bean(name = ["globalObjectMapper"])
-  @Primary
-  fun objectMapper(): ObjectMapper {
```
sp-watson, line 19: Has this been eplaced or is it not needed any more?!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/HmppsTierExceptionHandler.kt
```diff
@@ -0,0 +1,78 @@
...
+
+@RestControllerAdvice
+class HmppsTierExceptionHandler {
+
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @ExceptionHandler(EntityNotFoundException::class)
+  fun handleEntityNotFoundException(e: EntityNotFoundException): Mono<ResponseEntity<ErrorResponse>> {
```
sp-watson, line 24: It's mighty confusing that these exc handlers use Mono when we use coroutines everywhere... I suspect it's not necessary or there is a coroutine version?!

src/main/resources/logback-spring.xml
```diff
@@ -1,60 +1,47 @@
...
-
-    <logger name="org.springframework.boot" additivity="false" level="INFO">
-        <appender-ref ref="logAppender"/>
-        <appender-ref ref="consoleAppender"/>
-    </logger>
-
-    <root level="INFO">
-        <appender-ref ref="logAppender"/>
-    </root>
+<include resource="org/springframework/boot/logging/logback/defaults.xml"/>
```
sp-watson, line 3: We need to check App Insights actually works with all these logging and open telemetry changes

sp-watson (dismissed):
As with the other reactive PRs there are just too many changes to do a "proper" review, so i have tried to spot any things that don't seem to make sense to me - but they are probably OK.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
carlov20: The return type `[{...},{...}]` which is in effect a Flow, the one below is `{nsis:[{...},{...}]}` which is a single object wrapping an array so needs to be an awaitBody

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/ObjectMapperConfiguration.kt
carlov20: Spring initialises this through config https://github.com/ministryofjustice/hmpps-tier/blob/main/src/main/resources/application.yml#L7 so this wasn't needed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/client/CommunityApiClient.kt
sp-watson: Ok... oh-hum it just seems unnecessarily complicated (that's kotlin's approach not your code) - i think its trying to preserve interoperability with Java Reactive whilst adding primitives that make some things easier to do?!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/ObjectMapperConfiguration.kt
sp-watson: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/HmppsTierExceptionHandler.kt
carlov20: coroutines works

PR #439, opened 2023-05-12
WFP-1922: refactor ChangeLevelCalculator

AndySweilam:


carlov20 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/entity/TierCalculationEntity.kt
```diff
@@ -32,4 +33,19 @@ data class TierCalculationEntity(
   @Type(JsonType::class)
   @Column(columnDefinition = "jsonb")
   val data: TierCalculationResultEntity,
-)
+) {
+  override fun equals(other: Any?): Boolean {
```
carlov20, line 37: how come you had to do this?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/DeliusCommunityDataReliability.kt
```diff
@@ -71,37 +71,34 @@ class DeliusCommunityDataReliability(
...
+  @GetMapping("/crn/all/{offset}/{limit}")
+  suspend fun getAllDataReliability(@PathVariable(required = true) offset: Int, @PathVariable(required = true) limit: Int): Flow<CommunityDeliusData> {
+    return tierReader.getCrns(offset, limit).mapNotNull {
       val deliusInputs = try {
-        tierToDeliusApiService.getTierToDelius(it)
+        tierToDeliusApiClient.getDeliusTierTest(it)
       } catch (e: WebClientException) {
         log.error("Webclient exception in Tier To Delius for CRN: $it", e)
-        DeliusInputs(false, BigDecimal.valueOf(-1), -10, false, false, Registrations(false, emptyList(), null, null))
+        null
```
carlov20, line 81: how will you know if tier to delius returns a 404 and community API returns a 200 if you are filtering out all deliusInputs?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/repository/TierCalculationRepository.kt
```diff
@@ -13,6 +14,9 @@ interface TierCalculationRepository : CrudRepository<TierCalculationEntity, Long
 
   fun findByCrnAndUuid(crn: String, calculationId: UUID): TierCalculationEntity?
 
-  @Query("SELECT DISTINCT crn FROM TierCalculationEntity ORDER BY crn DESC")
-  fun findDistinctCrn(): List<String>
+  @Query(nativeQuery = true, value = "SELECT DISTINCT crn FROM tier_calculation ORDER BY crn DESC OFFSET ?1 LIMIT ?2")
```
carlov20, line 17: you could have used `Pageable` and return a `Page` here both things Spring Data has out of the box https://stackoverflow.com/questions/22345081/spring-data-jpa-query-and-pageable so that you can get the total pages to query

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/entity/TierCalculationEntityTest.kt
```diff
@@ -14,17 +14,22 @@ internal class TierCalculationEntityTest {
...
+    val tierCalculationResultEntity2 = TierCalculationEntity(id, calculationId, crn, created, data)
 
     assertThat(tierCalculationResultEntity.crn).isEqualTo(crn)
     assertThat(tierCalculationResultEntity.created).isEqualTo(created)
     assertThat(tierCalculationResultEntity.data).isEqualTo(data)
-
     assertThat(tierCalculationResultEntity.uuid).isEqualTo(calculationId)
+    assertThat(tierCalculationResultEntity.id).isEqualTo(id)
+    assertThat(tierCalculationResultEntity).isEqualTo(tierCalculationResultEntity2)
+    assertThat(tierCalculationResultEntity.toString()).isEqualTo(tierCalculationResultEntity2.toString())
```
carlov20, line 31: this test is now testing two things, if you wanted to test this write another test

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/entity/TierCalculationEntity.kt
AndySweilam: There were warnings about this [can cause severe performance and memory consumption issues]

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/repository/TierCalculationRepository.kt
AndySweilam: yes, I think having a limit/offset will give us more control on specific number of rows we want.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/DeliusCommunityDataReliability.kt
AndySweilam: yes, this will be a separate PR, at the moment I want to see how this will behave when deployed and if we can get rid of the time out errors. We will also need to add requirements.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/jpa/repository/TierCalculationRepository.kt
carlov20: you can supply a size with pageable and then page number does the offset, the comment was more of an FYI for the future when doing pagination

carlov20 (approved):

PR #525, opened 2024-01-03
PI-1757 updated to accept merge and gdpr events and move to virtual t…

anthony-britton-moj:
…hreads

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,161 +1,161 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.11.0"
-  kotlin("plugin.spring") version "1.9.21"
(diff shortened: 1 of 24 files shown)


build.gradle.kts
```diff
@@ -125,11 +111,11 @@ tasks {
             rule {
                 limit {
                     counter = "BRANCH"
-                    minimum = BigDecimal(0.88)
+                    minimum = BigDecimal(0.84)
                 }
                 limit {
                     counter = "COMPLEXITY"
-                    minimum = BigDecimal(0.88)
+                    minimum = BigDecimal(0.84)
```
marcus-bcl, line 118: Is there a reason for this number specifically ?

marcus-bcl (approved):
Looks good.  Getting rid of detekt means we can upgrade Kotlin etc too now right?

PR #528, opened 2024-01-05
PI-1747 added tier summary

anthony-britton-moj:


marcus-bcl (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
```diff
@@ -12,7 +12,7 @@ import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.hmppstier.dto.TierDto
 import uk.gov.justice.digital.hmpps.hmppstier.service.TierReader
 import uk.gov.justice.digital.hmpps.hmppstier.service.exception.EntityNotFoundException
-import java.util.UUID
+import java.util.*
```
pmcphee77, line 15: Do we want this wildcard import? A few below. Not sure if this is a coding standard here?

pmcphee77 (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/controller/TierCalculationController.kt
marcus-bcl: We recently switched from ktlint which enforces no wildcard imports to IntelliJ's code formatter which is a bit more lenient.  There's some reasoning in PI-1724.

PR #575, opened 2024-03-25
PI-2019 Update schedule for CP testing

marcus-bcl:
Depends on https://github.com/ministryofjustice/hmpps-helm-charts/pull/143

### DIFF
--- a/helm_deploy/hmpps-tier/values.yaml
+++ b/helm_deploy/hmpps-tier/values.yaml
@@ -26,6 +26,12 @@ generic-service:
     annotations:
       nginx.ingress.kubernetes.io/custom-http-errors: "418"
       external-dns.alpha.kubernetes.io/aws-weight: "100"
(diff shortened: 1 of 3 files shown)


helm_deploy/hmpps-tier/values.yaml
```diff
@@ -26,12 +26,6 @@ generic-service:
     annotations:
       nginx.ingress.kubernetes.io/custom-http-errors: "418"
       external-dns.alpha.kubernetes.io/aws-weight: "100"
-      nginx.ingress.kubernetes.io/configuration-snippet: |
```
petergphillips, line 29: can be replaced by `server-snippet` - see https://github.com/ministryofjustice/hmpps-helm-charts/releases/tag/generic-service-3.0.0

marcus-bcl: Oh nice, thanks.  Was just going to fall back to using the default role but I'll add that in.

